PR #6, opened 2019-04-29
Add Scheduler that handles the state machine of subscription fetching, charging and month validation

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +35,17 @@ data class Entry(
...
             addedBy = entryDto.addedBy ?: "system"
         )
+
+        fun from(memberSubscriptionDto: MemberSubscriptionDto): Entry = Entry(
+            id = UUID.randomUUID(),
+            memberId = memberSubscriptionDto.memberId,
+            fromDate = LocalDate.now(),
+            amount = memberSubscriptionDto.subscription,
+            type= EntryType.FEE,
+            source = "account-service",
```
palmfjord: i believe this should be `product-pricing`, no?

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +35,17 @@ data class Entry(
...
+            reference = LocalDate.now().toString(),
```
palmfjord: Also i think this should be the product id rather than the date

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +35,17 @@ data class Entry(
...
+            title = "TESTING",
```
palmfjord: why?

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +35,17 @@ data class Entry(
...
+            addedBy = "account-service"
```
palmfjord: `system` no?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.hedvig.accountservice.integration.productPricing.dto.MemberSubscriptionDto
+import org.springframework.cloud.openfeign.FeignClient
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestParam
+
+@FeignClient(
+    name = "product-pricing",
+    url = "\${hedvig.product-pricing.url:product-pricing}"
+    // configuration = FeignConfiguration::class.java TODO: Doesn't want to cooperate
```
palmfjord: 🙈

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
```diff
@@ -0,0 +1,20 @@
...
+)
+interface ProductPricingClient {
+
+    @GetMapping("/_/insurance/monthlyBilling")
+    fun getMonthlyBilling(
+       @RequestParam year: Int,
```
palmfjord: maybe we could make this  a `Year` type?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
```diff
@@ -0,0 +1,20 @@
...
+       @RequestParam month: Int
```
palmfjord: ...and this a `Month` type?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingServiceImpl.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
+import java.time.YearMonth
+
+@Service
+class ProductPricingServiceImpl @Autowired constructor(
+    private val productPricingClient: ProductPricingClient
+): ProductPricingService {
+    override fun getAllMemberSubscriptions(): List<MemberSubscriptionDto> {
+        val yearMonth = YearMonth.now()
```
palmfjord: i think this should be an argument rather than a hard-coded `now`

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingServiceImpl.kt
```diff
@@ -0,0 +1,16 @@
...
+        return this.productPricingClient.getMonthlyBilling(yearMonth.year, yearMonth.monthValue)
```
palmfjord: do we need `this`?

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/scheduling")
+class SchedulingController @Autowired constructor(
+    private val productPricingService: ProductPricingService,
+    private val commandGateway: CommandGateway
+    ) {
+
+    @PostMapping("/addFeeEntryToAllMembers")
```
palmfjord: I know we're not very consistent about it, but what's your thoughts on camel vs kebab case in urls? I'm much more a fan of kebab

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun addFeeEntryToAllMembers(): ResponseEntity<Void> {
+        val allMembersSubscriptions = productPricingService.getAllMemberSubscriptions()
+        allMembersSubscriptions.forEach {
+            val entry = Entry.from(it)
```
palmfjord: can we avoid `it` and make a named parameter to this lambda? I think `it` gets pretty hard to read

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
vonElfvin: Not in product-pricing, would it be better to keep it as a `Month` to make it more foolproof?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingServiceImpl.kt
vonElfvin: Agreed

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingServiceImpl.kt
vonElfvin: Don't think so, good catch

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
vonElfvin: I saw camel in both product and payment service. Using a single word is my personal preference 😭

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: Probably yes!

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: I fully agree, if that's the case we need to make our own endpoint in product-pricing. Since the one we use now "only" returns memberId + subscription

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
palmfjord: Hm i'm pretty sure they translate into numbers in the json anyways, so it _shouldn't_ matter and i like the explicity

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
palmfjord: haha! maybe stick to camel then even though it hurts

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: Opinions on how to handle this?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
vonElfvin: One error that I've experienced now is that having month as month type casts the number 4 to month of May since Javascript calculates months from 0. Should I still use month type? I'm considering changing my own endpoint to Int just to avoid this confusion.

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
palmfjord: I think it could be worth either modifying the endpoint or making a new similar one :) i think we'd shoot ourselves in the foot not being able to track what we're actually charging for

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
palmfjord: I love time 😍 idk, in a way maybe its better to just keep this as one `YearMonth` instead, then we don't have any ambiguity as it'll be a string like `2019-05`

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -19,6 +22,7 @@ data class Entry(
     val comment: String? = null,
     val addedBy: String = "system",
     val addedAt: Instant = Instant.now()
+    // TODO: Add a "chargeId" to keep track of what charge that payed for the entry
```
palmfjord: Why can't we use `reference`?

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +38,29 @@ data class Entry(
...
+
+        fun from(event: ScheduleSubscriptionEvent): Entry = Entry(
+            id = UUID.randomUUID(),
+            memberId = event.memberId,
+            fromDate = LocalDate.of(event.year.value, event.month, 1),
+            amount = event.subscription,
+            type= EntryType.FEE,
+            source = "product-pricing",
+            reference = LocalDate.now().toString(),
+            title = "SCHEDULED AUTOMATIC FEE",
```
palmfjord: Do we need this? I think it's better to leave it as `null` and compute this jit

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +38,29 @@ data class Entry(
...
+        )
+
+        fun from(event: CompleteChargeEvent, chargedBy: String?): Entry = Entry(
+            id = UUID.randomUUID(),
+            memberId = event.memberId,
+            fromDate = LocalDate.ofInstant(event.chargedAt, ZoneOffset.UTC),
+            amount = event.amount.negate(),
+            type= EntryType.CHARGE,
+            source = "payment-service",
+            reference = LocalDate.now().toString(),
```
palmfjord: same, i think this should be the tx id or something rather than the date :)

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +38,29 @@ data class Entry(
...
+            title = "SCHEDULED AUTOMATIC CHARGE",
```
palmfjord: and same :D Jit computation

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
```diff
@@ -14,35 +14,57 @@ interface EntryRepository {
...
         SELECT COALESCE(SUM(amount), 0)
             FROM "account_entries"
             WHERE "member_id" = :memberId
-            AND "from_date" <= :today
-        ;
+            AND "from_date" <= :today;
         """
     )
-    fun sumCurrentMonth(@Bind("memberId") memberId: String, @Bind("today") today: LocalDate = YearMonth.now().atEndOfMonth()): MonetaryAmount
+    fun sumCurrentMonth(
```
palmfjord: Maybe this should be `sumUntil` rather than something about month? Considering it's actually summing until a certain date

src/main/kotlin/com/hedvig/accountservice/schedules/domain/sagas/ChargeScheduledSaga.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.axonframework.eventhandling.saga.SagaLifecycle
+import org.axonframework.eventhandling.saga.StartSaga
+import org.axonframework.spring.stereotype.Saga
+import org.springframework.beans.factory.annotation.Autowired
+
+@Saga
+class ChargeScheduledSaga {
+
+    private val logger = KotlinLogging.logger { }
+    private var scheduledBy: String? = null
```
palmfjord: are we sure this is ok? This feels **very** scary from a concurrency perspective if it doesn't instantiate a new saga for every use

src/main/kotlin/com/hedvig/accountservice/schedules/domain/sagas/ChargeScheduledSaga.kt
```diff
@@ -0,0 +1,93 @@
...
+                chargedAt = event.chargedAt,
+                chargedBy = scheduledBy!!,
+                transactionId = event.transactionId
+            ))
+        }
+    }
+
+    @SagaEventHandler(associationProperty = "transactionId")
+    @EndSaga
+    fun on(event: ChargeFailedEvent) { }
```
palmfjord: nothing?

src/main/kotlin/com/hedvig/accountservice/schedules/model/SchedulerStateManager.kt
```diff
@@ -0,0 +1,30 @@
+package com.hedvig.accountservice.schedules.model
+
+import org.jdbi.v3.core.Jdbi
+import org.springframework.stereotype.Component
+
+@Component
+class SchedulerStateManager(
+    private val jdbi: Jdbi
+) {
+    /*fun findSchedulerStates(year: Year, month: Month, status: ChargeStatus?): List<SchedulerState> {
```
palmfjord: 😅

src/main/kotlin/com/hedvig/accountservice/schedules/model/SchedulerStateRepository.kt
```diff
@@ -0,0 +1,50 @@
...
+        """
+    )
+    fun findAllLatest(): List<SchedulerState>
+
+    @SqlUpdate(
+        """
+
+        """
+    )
+    fun addSubscriptionScheduled(
```
palmfjord: no?

src/main/kotlin/com/hedvig/accountservice/schedules/model/SubscriptionScheduledRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+    )
+    fun find(
+        @Bind memberId: String,
+        @BindBean("year") year: Year,
+        @BindBean("month") month: Month
+    ): SubscriptionScheduled
+
+    @SqlQuery(
+        """
+            SELECT EXISTS(
```
palmfjord: NEAT

src/main/kotlin/com/hedvig/accountservice/schedules/model/SubscriptionScheduledRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+            SELECT *
+                FROM "subscription_scheduled"
+                WHERE member_id=:memberId AND year=:year.value AND month=:month.value
+        """
+    )
+    fun find(
+        @Bind memberId: String,
+        @BindBean("year") year: Year,
+        @BindBean("month") month: Month
+    ): SubscriptionScheduled
```
palmfjord: maybe this should actually be nullable no?

src/main/kotlin/com/hedvig/accountservice/schedules/web/ScheduleController.kt
```diff
@@ -0,0 +1,147 @@
...
+    private val stateManager: SchedulerStateManager,
+    private val productPricingService: ProductPricingService
+    ) {
+    @PostMapping("/reset/{memberId}")
+    fun resetMember(
+        @PathVariable memberId: String,
+        @RequestParam resetBy: String
+    ): ResponseEntity<Void> {
+        commandGateway.sendAndWait<ResetSchedulerCommand>(ResetSchedulerCommand(memberId, resetBy))
+        return ResponseEntity.ok().build()
```
palmfjord: maybe `ResponseEntity.noContent()` is better

palmfjord: (applies to all ofc)

src/main/kotlin/com/hedvig/accountservice/schedules/web/ScheduleController.kt
```diff
@@ -0,0 +1,147 @@
...
+        }
+        return ResponseEntity.ok().build()
+    }
+
+    @PostMapping("/charge/{memberId}")
+    fun chargeMember(
+        @PathVariable memberId: String,
+        @RequestParam scheduledBy: String
+    ): ResponseEntity<Void> {
+        val account = accountManager.findAccount(memberId)
```
palmfjord: This does not feel safe. It should either be done in a single db tx imo, or it should be sent from the frontend so we can verify the intended amount is the same as the amount that'll be charged

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
```diff
@@ -14,35 +14,74 @@ interface EntryRepository {
...
+            UPDATE "account_entries"
+                SET type=:type
+                WHERE id=:entryId
+        """
+    )
+    fun updateType(@Bind entryId: UUID, @Bind type: EntryType)
+
+    @SqlUpdate(
+        """
+            DELETE FROM "account_entries"
```
palmfjord: We should never delete things imo, but rather timestamp them as `deletedAt` or similar

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
```diff
@@ -14,35 +14,74 @@ interface EntryRepository {
...
+            FROM "account_entries"
+            WHERE "from_date" <= :today
+            GROUP BY member_id;
+        """
+    )
+    fun sumAllCurrentMonth(@Bind("today") today: LocalDate = YearMonth.now().atEndOfMonth()): List<AccountBalance>
+
+    @SqlUpdate(
+        """
+            UPDATE "account_entries"
```
palmfjord: also we should never update things, but rather make add things to some kind of separate "history" table and pick the latest.

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -53,22 +47,22 @@ class MemberAccount() {
...
-        AggregateLifecycle.apply(EntryDeletedEvent(command.memberId, command.entryId))
+    fun handle(command: FailChargeEntryCommand) {
+        AggregateLifecycle.apply(ChargeEntryFailedEvent(command.memberId, command.entryId, command.failedAt))
     }
 
     @EventSourcingHandler
-    fun on(event: EntryDeletedEvent) {
-        entries.removeIf { entry -> entry.id == event.entryId }
+    fun on(event: ChargeEntryFailedEvent) {
+        entries.map { entry -> if (entry.id == event.entryId) entry.failedAt = event.failedAt else entry }
```
palmenhq: can we do brackets around ifs? Even if this is a oneliner it's so easy to miss and make fallthroughs

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -70,13 +71,13 @@ class ScheduleController @Autowired constructor(
     fun addSubscriptionToMember(
         @PathVariable memberId: String,
         @RequestParam scheduledBy: String
-    ): ResponseEntity<Void> {
+    ): ResponseEntity<String> {
```
palmenhq: i think this should be some kind of object, not just a pure `String`

src/main/resources/db/changelog/changelog-master.yml
```diff
@@ -1,6 +1,6 @@
 databaseChangeLog:
   - include:
-      file: ./migrations/20190418-1340-create-account_entries.sql
+      file: ./migrations/20190520-1040-create-account_entries.sql
```
palmenhq: this should come after https://github.com/HedvigInsurance/account-service/pull/6/files/522f41925812927753391afe94758316926f6ce2..96cc6ad354f5f51ae4f34d6c63d1ebfc69064a0d#diff-69bcfc2273b3e6caefdd21ae5dbaae32R6 if we change the date to be after

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
vonElfvin: I was thinking of this more as a ternary than an if statement. I personally think this is more readable given how ternary operators usually look like?

src/main/resources/db/changelog/changelog-master.yml
vonElfvin: Could you clarify what you mean with "after"? 🙈

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
palmenhq: sure, sounds reasonable 👍

src/main/resources/db/changelog/changelog-master.yml
palmenhq: It should come last in the file if its last in the file tree also. But I don't really see why  we would want to move this in the first place

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
```diff
@@ -62,7 +77,8 @@ interface EntryRepository {
         """
         SELECT member_id, COALESCE(SUM(amount), 0) AS current_balance
             FROM "account_entries"
-            WHERE "from_date" <= :today
+            WHERE from_date <= :today
+                AND (failed_at IS NULL OR charged_at IS NOT NULL)
```
palmenhq: I don't understand exactly how this works, feels a bit ambiguous don't you think? Will it pick `failed_at` or `charged_at`? idk 🤷‍♀️

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
palmenhq: Actually, coming back to this.. 😬 this should probably be a `foreach` rather than a `map` considering we're not returning anything  (applies to the other one as well) and hence we shouldn't use a ternary in the first place

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +42,38 @@ data class Entry(
             comment = entryDto.comment,
             addedBy = entryDto.addedBy ?: "system"
         )
+
+        fun from(event: ScheduleSubscriptionEvent, period: YearMonth, productSubscription: InsuranceProductSubscriptionDto): Entry = Entry(
+            id = UUID.randomUUID(),
+            memberId = event.memberId,
+            fromDate = LocalDate.of(period.year, period.month, 1),
```
palmenhq: Just for future reference we could also do `period.atDay(1)` :)

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -34,5 +42,38 @@ data class Entry(
...
+            type = EntryType.SUBSCRIPTION,
+            source = "product-pricing",
+            reference = productSubscription.productId.toString(),
+            addedBy = event.scheduledBy
+        )
+
+        fun from(event: ChargeRequestedEvent, entryId: UUID, transactionId: UUID): Entry = Entry(
+            id = entryId,
+            memberId = event.memberId,
+            fromDate = LocalDate.now(),
```
palmenhq: No, we should base the `LocalDate` on the timestamp of the event rather than `now` as the events _may_ be replayed and hence have a different real date

palmenhq: And i guess we should pass the timestamp on to the `addedAt` prop no?

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        val previousPeriod = period.minusMonths(1)
+
+        if (shouldCheckPeriod(previousPeriod, memberId, checkAllPeriods)) {
+            checkSubscriptionsOfPeriod(event, previousPeriod, memberId, checkAllPeriods)
+        }
+    }
+
+    private fun shouldCheckPeriod(period: YearMonth, memberId: String, checkAllPeriods: Boolean = false): Boolean {
```
palmenhq: very nice

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SubscriptionScheduledRepository.kt
```diff
@@ -0,0 +1,35 @@
...
+        """
+    )
+    fun add(@BindBean subscriptionScheduled: SubscriptionScheduled)
+
+    @SqlQuery(
+        """
+            SELECT EXISTS(
+                SELECT 1
+                    FROM "subscription_scheduled"
+                    WHERE member_id=:memberId AND year=:year.value AND month=:month.value
```
palmenhq: have we not fixed the `(Year|Month)Argument` in JDBI?

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -0,0 +1,161 @@
...
+        }
+        return ResponseEntity.noContent().build()
+    }
+
+    @PostMapping("/approvals")
+    fun approveChargeAmount(
+        @RequestBody requests: List<ApproveChargeRequestDto>,
+        @RequestParam approvedBy: String
+    ): ResponseEntity<String> {
+        if (requests.distinctBy { request -> request.memberId }.size != requests.size) {
```
palmenhq: boom, really neat.

src/main/resources/db/changelog/changelog-master.yml
vonElfvin: image
Like this?

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
vonElfvin: It ignores failed charges (they failed_at=NULL)  unless they have been completed later (charged_at != NULL)

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SubscriptionScheduledRepository.kt
vonElfvin: Unsure if it works for queries as well? Will check.

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: Good catch, this needs to be changed in previous Entry.from as well. IMO Entry's addetAt should not have a default value to remove this mistake in the future.

src/main/resources/db/changelog/changelog-master.yml
palmenhq: Yes 🙌

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
palmenhq: Yes, but I get the feeling it's a bit implicit don't you think? I'm slightly unsure exactly how to make it more explicit and clear this is the behaviour we want but I would not understand that's what we want to achieve when coming back to this later

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
palmenhq: Yes, that sounds very good!

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SubscriptionScheduledRepository.kt
palmenhq: Ah, I guess it should from the `(Year|Month)Mapper`

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: Because let's say the reference is the productId then we can't change it

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
vonElfvin: Given that it works, only used for tests and is pretty neat I'd say its good enough 🤔

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SubscriptionScheduledRepository.kt
vonElfvin: It worked when I changed it from @BindBean to @Bind 😄

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: Given that "Entry.from()" is only called from Controllers and Sagas my verdict is that this is fine. The entry will be inside the event being replayed and thus have the correct date even if replayed.

src/test/kotlin/com/hedvig/accountservice/schedulers/domain/MemberSchedulerTest.kt
```diff
@@ -0,0 +1,821 @@
...
+                assertThat(account.scheduler.currentState.memberId).isEqualTo("123")
+                assertThat(account.scheduler.currentState.status).isEqualTo(ChargeStatus.SUBSCRIPTION_SCHEDULED_AND_WAITING_FOR_APPROVAL)
+                assertThat(account.scheduler.currentState.changedBy).isEqualTo("subscription@hedvig.com")
+                assertThat(account.scheduler.currentState.amount).isNull()
+                assertThat(account.scheduler.currentState.transactionId).isNull()
+            }
+    }
+
+    @Test
+    fun schedulesSubscriptionWhenSubscriptionAlreadyScheduled() {
```
palmenhq: i feel like this test name is not true, `doesntScheduleSubscriptionWhenSubscriptionAlreadyScheduled` would be more appropriate no?

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
vonElfvin: I made a new endpoint in product-pricing 👍

src/test/kotlin/com/hedvig/accountservice/schedulers/domain/MemberSchedulerTest.kt
vonElfvin: Nice catch!

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -0,0 +1,171 @@
...
+        return ResponseEntity.noContent().build()
+    }
+
+    @PostMapping("/charges")
+    fun chargeAllMembers(
+        @RequestParam scheduledBy: String
+    ): ResponseEntity<Void> {
+        val approvedMemberIds = stateManager.findAllMemberIds(ChargeStatus.APPROVED_FOR_CHARGE)
+        approvedMemberIds.forEach { memberId ->
+            val transition = commandGateway.sendAndWait<SchedulerStateTransition>(ScheduleChargeCommand(
```
palmenhq: I guess there's a risk this will DoS payment-service or Trustly as sagas and `@EventHandler`s are asynchronous, meaning `sendAndWait` will actually not wait  for each charge to continue

palmenhq: Lets investigate and see if we need

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/ChargeScheduledSaga.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.axonframework.eventhandling.saga.StartSaga
+import org.axonframework.spring.stereotype.Saga
+import org.springframework.beans.factory.annotation.Autowired
+import java.time.Instant
+import java.util.*
+
+@Saga(sagaStore = "jdbcSagaStore")
+class ChargeScheduledSaga {
+
+    private val logger = KotlinLogging.logger{}
```
palmenhq: should this not be `@Transient`?

palmenhq: (all loggers)

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/ChargeScheduledSaga.kt
```diff
@@ -0,0 +1,144 @@
...
+        logger.info { "Charge entry updated to COMPLETED for memberId=${event.memberId} with transactionId=${event.transactionId} and entryId=$entryId" }
+    }
+
+    @SagaEventHandler(associationProperty = "transactionId")
+    @EndSaga
+    fun on(event: ChargeFailedEvent) {
+        commandGateway.sendAndWait<Void>(FailChargeEntryCommand(
+            memberId = event.memberId,
+            entryId = entryId!!,
+            failedAt = Instant.now()
```
palmenhq: should this not be a timestamp from `on(event: ChargeFailedEvent, @Timestamp now: Instant)`?

src/main/kotlin/com/hedvig/accountservice/accounts/web/AccountController.kt
```diff
@@ -34,7 +39,12 @@ class AccountController @Autowired constructor(
         @RequestBody entryDto: EntryRequestDto
     ): ResponseEntity<UUID> {
         val entry = Entry.from(entryDto, memberId)
-        commandGateway.sendAndWait<AddEntryCommand>(AddEntryCommand(memberId, entry))
+
+        if (entry.fromDate.isBefore(LocalDate.now())) {
+            return ResponseEntity.status(HttpStatus.FORBIDDEN).build()
```
palmenhq: still forbidden

PR #7, opened 2019-04-30
WIP Add endpoint that charges member(s) what the owe

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryType.kt
```diff
@@ -4,5 +4,5 @@ enum class EntryType {
     CORRECTION,
     FEE,
     CAMPAIGN,
-    PAYMENT
+    PAYMENT // TODO: Split up to PAYOUT and CHARGE?
```
palmfjord: should we even include `payout`?

palmfjord: i like `charge` tho

src/main/kotlin/com/hedvig/accountservice/integration/paymentService/PaymentServiceClient.kt
```diff
@@ -0,0 +1,20 @@
...
+@FeignClient(
+    name = "payment-service",
+    url = "\${hedvig.payment-service.url:payment-service}"
+)
+interface PaymentServiceClient {
+    @PostMapping("/_/members/{memberId}/charge")
+    fun chargeMember(
+        @PathVariable("memberId") memberId: String,
+        @RequestBody chargeRequestDto: ChargeRequestDto
+    ): ResponseEntity<Any>
```
palmfjord: `Any` :/

src/main/kotlin/com/hedvig/accountservice/integration/paymentService/PaymentServiceImpl.kt
```diff
@@ -0,0 +1,19 @@
...
+import javax.money.MonetaryAmount
+
+@Service
+class PaymentServiceImpl @Autowired constructor(
+    private val paymentServiceClient: PaymentServiceClient
+): PaymentService {
+    override fun chargeMember(memberId: String, amount: MonetaryAmount): ResponseEntity<ChargeResponseDto> {
+        val response = paymentServiceClient.chargeMember(memberId, ChargeRequestDto(amount))
+        val body = ChargeResponseDto(memberId, amount) // TODO: Discuss
+        return ResponseEntity.status(response.statusCode).headers(response.headers).body(body)
```
palmfjord: why don't we just pass on the whole response entity?

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountBalance.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.accountservice.accounts.model
+
+import javax.money.MonetaryAmount
+
+data class AccountBalance (val memberId: String, val currentMonthsBalance: MonetaryAmount)
```
palmfjord: Just some nitty gritty stuff but I believe it's more common not to have this space between the class name and the first parenthesis

palmfjord: (i like how it's very consistent though, but hence it applies to all classes)

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
```diff
@@ -29,6 +31,37 @@ class SchedulingController @Autowired constructor(
         return ResponseEntity.ok().build()
     }
 
+    // TODO: Doesn't work since aggregates don't exist on payment service, handle error?
+    @PostMapping("/charge/all")
+    fun chargeAllMembers(): ResponseEntity<Void> {
+        val allAccountBalances = accountManager.allAccountBalances()
+        allAccountBalances.forEach {
+            val response = paymentService.chargeMember(it.memberId, it.currentMonthsBalance)
```
palmfjord: i'd prefer to use an explicit name rather than `it` :) as it's pretty hard to know what `it` is sometimes

src/test/kotlin/com/hedvig/accountservice/accounts/model/EntryRepositoryTest.kt
```diff
@@ -191,4 +191,87 @@ class EntryRepositoryTest {
...
+            entryRepository.add(entry4)
+            entryRepository.add(entry5)
+            entryRepository.sumAllCurrentMonth()
+        }
+
+        assertThat(result).hasSize(2)
+        assertThat(result).isNotNull
+        assertThat(result[0].memberId).isEqualTo(entry2.memberId)
+        assertThat(result[1].memberId).isEqualTo(entry1.memberId)
+        assertThat(result[0].currentMonthsBalance).isEqualTo(Money.of(BigDecimal.valueOf(1337), "SEK"))
```
palmfjord: should we maybe assert the member ids also? so like we know that `1337` is for the correct member?

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
```diff
@@ -14,35 +14,57 @@ interface EntryRepository {
...
+
+    @SqlQuery(
+        """
+        SELECT member_id, COALESCE(SUM(amount), 0) AS current_months_balance
+            FROM "account_entries"
+            WHERE "from_date" <= :today
+            GROUP BY member_id;
+        """
+    )
+    fun sumAllCurrentMonth(@Bind("today") today: LocalDate = YearMonth.now().atEndOfMonth()): List<AccountBalance>
```
palmfjord: I'm slightly unsure about this but maybe we could return a `Map<String, MonetaryAmount>`? _should_ be possible ™️

src/test/kotlin/com/hedvig/accountservice/accounts/model/EntryRepositoryTest.kt
vonElfvin: Is that not what happens at lin 272+273? Or am I missing something

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
vonElfvin: I considered that but wanted to a have Entry.from() using Map<String, Monetary> amount then. Figured from using AccountBalance was more "powerful" in the long run. Thoughts?

src/main/kotlin/com/hedvig/accountservice/integration/paymentService/PaymentServiceClient.kt
vonElfvin: The issue is that payment-service returns ResponseEntity\<?\>. In practise it returns an empty string "". Change to ResponseEntity(String) maybe?

src/main/kotlin/com/hedvig/accountservice/integration/paymentService/PaymentServiceImpl.kt
vonElfvin: Because I wanted to append the successful charge, the information is however available outside the method. We need to discuss the whole flow because now I belive it is not that great.

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryType.kt
vonElfvin: I understand it as Payment going both ways (either a payout or a charge). Hence the word payout to make it explicit.

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
vonElfvin: I was thinking that maybe we should use something else than a forEach loop here regardless. Could not come up with anything better at the time. Any ideas of improvements?

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryRepository.kt
palmfjord: Fair enough! Or maybe `Map<String, AcocuntBalance>`?

src/test/kotlin/com/hedvig/accountservice/accounts/model/EntryRepositoryTest.kt
palmfjord: yes it is... plz ignore me

src/main/kotlin/com/hedvig/accountservice/integration/paymentService/PaymentServiceClient.kt
palmfjord: Ah, or just `ResponseEntity<Void>`? So we discard this knowledge of an empty string?

src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryType.kt
palmfjord: Fair enough! go ahead :D

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
palmfjord: Yeah that's fair. coming back to this i think we should actually make a bulk endpoint in `paymentService` rather than making several thousand calls to in

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
```diff
@@ -29,6 +31,37 @@ class SchedulingController @Autowired constructor(
         return ResponseEntity.ok().build()
     }
 
+    // TODO: Doesn't work since aggregates don't exist on payment service, handle error?
+    @PostMapping("/charge/all")
```
palmfjord: doesn't this collide with `/charge/{memberId}`?

src/main/kotlin/com/hedvig/accountservice/scheduling/web/SchedulingController.kt
vonElfvin: Exactly 🙈

PR #9, opened 2019-05-27
Add limit request param to listStates endpoint

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/schedulers/model/SchedulerStateRepository.kt
```diff
@@ -20,10 +20,11 @@ interface SchedulerStateRepository {
         """
             SELECT DISTINCT ON (member_id) *
                 FROM "scheduler_states"
-                ORDER BY member_id, changed_at DESC;
+                ORDER BY member_id, changed_at DESC
+                LIMIT :limit;
```
palmenhq: what happens if its null?

vonElfvin: Then there is no limt = all are returned

palmenhq: ooh cool

PR #19, opened 2019-05-29
Scheduler denies charges of 10 sek or less

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -202,11 +206,26 @@ class MemberScheduler {
             logger.error { "Charged approval DENIED for memberId=${command.memberId} (Message: ${transition.message})" }
             return transition
         }
+        if (command.amount.isLessThanOrEqualTo(Money.of(BigDecimal.ZERO, "SEK"))) {
```
palmenhq: Are we sure we shouldn’t add some wiggle room, so we don’t charge like plus 5 or 10, also?

vonElfvin: Changed it to 10 sek

PR #20, opened 2019-06-03
Feature/deny same period reset

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -27,7 +27,7 @@ import java.time.ZoneId
 @Saga(sagaStore = "jdbcSagaStore")
 class SubscriptionScheduledSaga {
 
-    private val firstPeriod = YearMonth.of(2019, 5)
+    private val firstPeriod = YearMonth.of(2019, 6) // TODO: Change to 2018-02
```
palmenhq: huh? Why?

src/test/kotlin/com/hedvig/accountservice/schedulers/domain/MemberSchedulerTest.kt
```diff
@@ -725,6 +728,10 @@ class MemberSchedulerTest {
                 chargeInitiatedCommand,
                 chargeCompletedCommand
             )
+
+        mockkStatic(YearMonth::class)
+        every { YearMonth.now() } returns now.plusMonths(1)
```
palmenhq: please don't mock static methods, we've had so many issues with it already. If there's no other way then i think we should inject a getter lambda

palmenhq: this is even though we unmock it, just to be on the safe side

meletios:
@vonElfvin @palmenhq is this still relevant? Can you close it if not?

PR #23, opened 2019-07-29
Feature/rebased blacklist

vonElfvin:
Rebased my old blacklist feature to have the campaign changes. The rest is the same.


src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -109,7 +109,7 @@ class ScheduleController @Autowired constructor(
                 logger.error { transition.message }
             }
         }
-        Thread.sleep(250)
+        Thread.sleep(500)
```
palmenhq: i understand that we need to but it makes me sad

vonElfvin: Same

PR #24, opened 2019-08-16
Allow for attempting to automatically approve of normal cases of charges

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -128,6 +133,31 @@ class ScheduleController @Autowired constructor(
...
+    @PostMapping("/approvals/auto")
+    fun attemptAutoApproveChargeAmounts(
+        @RequestParam approvedBy: String
+    ): ResponseEntity<Void> {
+        val memberIdsWaitingForApproval = stateManager.findAllMemberIds(ChargeStatus.SUBSCRIPTION_SCHEDULED_AND_WAITING_FOR_APPROVAL)
+        val currentPeriod = YearMonth.now()
+        memberIdsWaitingForApproval.forEach { memberId ->
+            val monthlyBilling = productPricingService.getMonthlyBilling(memberId, currentPeriod)
+            val accountBalance = accountManager.getCurrentBalance(memberId)
+            if (accountBalance.isEqualTo(monthlyBilling.subscription)) {
```
meletios: Do we need to take into account the discounts here?

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -128,6 +133,31 @@ class ScheduleController @Autowired constructor(
         return ResponseEntity.noContent().build()
     }
 
+    @PostMapping("/approvals/auto")
+    fun attemptAutoApproveChargeAmounts(
+        @RequestParam approvedBy: String
+    ): ResponseEntity<Void> {
+        val memberIdsWaitingForApproval = stateManager.findAllMemberIds(ChargeStatus.SUBSCRIPTION_SCHEDULED_AND_WAITING_FOR_APPROVAL)
```
meletios: is it better to create a specific method without passing the enum? 

Something like this

```suggestion
        val memberIdsWaitingForApproval = stateManager.findAllMemberIdsWaitingForApproval()
```

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
vonElfvin: The idea is to only automatically approve the members who don't have any discounts, any object etc. Just plain old home insurance.

pom.xml
```diff
@@ -26,7 +26,7 @@
         <jdk>1.8</jdk>
         <kotlin.version>1.3.30</kotlin.version>
         <kotlin.code.style>official</kotlin.code.style>
-        <jackson.version>2.9.9</jackson.version>
+        <jackson.version>2.10.0.pr1</jackson.version>
```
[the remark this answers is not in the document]
vonElfvin: Because snyk is whining about our jackson version, the only one that satisfies the security is this unfortunally

PR #25, opened 2019-08-17
Feature/graphql

vonElfvin:


src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,31 @@
+schema {
```
hedvigoscar: @Iamsamwhoami @FredrikAreschoug 

I'd like you to check this schema out as well 🙂

src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,31 @@
...
+    query: Query
+    mutation: Mutation
+}
+
+type Query {
+    chargeEstimation(memberId: String): ChargeEstimation,
+    chargeHistory(memberId: String): [Charge!]!
```
Iamsamwhoami: `memberId` should be resolved from the token in the headers, not passed in by the consumer :)

src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,31 @@
...
+type Mutation {
+}
+
+type Charge {
+    amount: MonetaryAmountV2!,
+    date: LocalDate!
+}
+
+type ChargeEstimation {
+    currentBalance: MonetaryAmountV2!,
```
Iamsamwhoami: As this is not an estimation (i assume) maybe it shouldn't be inside this type?

src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,31 @@
...
+    estimatedSubscription: MonetaryAmountV2!,
```
Iamsamwhoami: it feels a bit redundant to prefix these with `estimated`

src/main/kotlin/com/hedvig/accountservice/graphql/extensions/DataFetchingEnviromentExtensions.kt
```diff
@@ -0,0 +1,15 @@
+package com.hedvig.accountservice.graphql.extensions
+
+import graphql.schema.DataFetchingEnvironment
+import graphql.servlet.GraphQLContext
+
+private const val HEDVIG_TOKEN: String = "hedvig.token"
+
+fun DataFetchingEnvironment.getToken() =
```
hedvigoscar: Actually so good

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
```diff
@@ -0,0 +1,91 @@
...
+        val memberId = env.getToken()
+        val currentBalance = accountManager.getCurrentBalance(memberId)
+
+        val estimatedSubscription = getEstimatedSubscription(memberId)
+        val estimatedDiscount = getEstimatedDiscount(memberId)
+
+        val failedCharges = if (currentBalance > estimatedSubscription.subtract(estimatedDiscount)) {
+            val chargeStates = stateManager.findAllChargeStates(memberId)
+            chargeStates
+                .subList(chargeStates.indexOfLast { it.status == ChargeStatus.CHARGE_COMPLETED }.coerceAtLeast(0), chargeStates.size)
```
vonElfvin: Does this fallback to index 0 or the last index when no completed charge is available? I.e. if we've never been able to charge a member ever.

pom.xml
```diff
@@ -272,6 +262,35 @@
...
+            <version>5.9.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.graphql-java-kickstart</groupId>
+            <artifactId>graphiql-spring-boot-starter</artifactId>
+            <version>5.9.0</version>
+        </dependency>
+
+        <!--- GraphQl Dependencies -->
+<!--        <dependency>-->
```
meletios: do we need the commented lines?

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SchedulerStateRepository.kt
```diff
@@ -61,4 +61,13 @@ interface SchedulerStateRepository {
...
     fun findAllLatestMemberIdsOfStatus(@Bind status: ChargeStatus): List<String>
+
+    @SqlQuery(
+        """
+            SELECT *
+            FROM "scheduler_states"
+            WHERE member_id=:memberId
+        """
+    )
+    fun findAllChargeStateOfMember(@Bind memberId: String): List<SchedulerState>
```
meletios: Should we include all the charges states here? blacklisted? or waiting for approval? 

would it be better to show only     
   CHARGE_REQUESTING,
    CHARGE_REQUEST_FAILED,
    CHARGE_INITIATED,
    CHARGE_FAILED,
    CHARGE_COMPLETED,

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Mutation.kt
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.accountservice.graphql.resolvers
+
+import com.coxautodev.graphql.tools.GraphQLMutationResolver
+import org.springframework.stereotype.Component
+
+@Component
+class Mutation: GraphQLMutationResolver
```
meletios: we can remove this i think since we dont have a mutation

PR #28, opened 2019-09-25
Estimation for next month

FredrikAreschoug:


src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
```diff
@@ -49,7 +50,6 @@ class Query(
 
     fun chargeEstimation(env: DataFetchingEnvironment): ChargeEstimation {
         val memberId = env.getToken()
-        // TODO: Return null when >= 27th and before the 1st?
```
palmenhq: nice catch

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
```diff
@@ -74,18 +74,19 @@ class Query(
...
     private fun getEstimatedDiscount(memberId: String): MonetaryAmount {
-        val now = YearMonth.now()
-        val currentDiscounts = productPricingService.getMonthlyCampaignDiscounts(memberId, now)
+        val currentDiscounts = productPricingService.getMonthlyCampaignDiscounts(memberId, getMonthForEstimation())
         return currentDiscounts
             .map { discount -> discount.amount }
             .fold(Money.of(0, "SEK")) { total, discount -> total.add(discount) }
     }
+
+    private fun getMonthForEstimation() =
```
palmenhq: idk if i like putting this logic in the query resolver but i guess its ok

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
vonElfvin: This comment was placed to remind us of the issue this PR is here to fix 😇

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
```diff
@@ -74,18 +74,19 @@ class Query(
     }
 
     private fun getEstimatedSubscription(memberId: String): MonetaryAmount {
-        val now = YearMonth.now()
-        val currentProductSubscriptions = productPricingService.getProductSubscriptions(memberId, now)
+        val currentProductSubscriptions = productPricingService.getProductSubscriptions(memberId, getMonthForEstimation())
```
vonElfvin: I'm personally a fan of putting variables outside function calls i.e. 
```suggestion
        val monthForEstimation = getMonthForEstimation()
        val currentProductSubscriptions = productPricingService.getProductSubscriptions(memberId, monthForEstimation)
```
to make the code clear. This is just an opinion though so it's nothing worth fixing.

src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
palmenhq: Yes that I understand, but i would've forgotten to remove it haha 😄

PR #33, opened 2019-10-03
Add reset handler for number of failed charges

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesController.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/numberFailedCharges")
+class NumberFailedChargesController @Autowired constructor(
+    private val eventProcessingConfiguration: EventProcessingConfiguration
+) {
+    @PostMapping("dangerouslyReset")
```
meletios:
```suggestion
    @DeleteMapping("dangerouslyReset")
```

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -15,49 +16,62 @@ import java.time.Instant
...
-        logger.info { "Outside charge completed: " +
-                "Failed charge number reset for member=${counter.memberId} to count=${counter.numberFailedCharges}" }
+        logger.info {
+            "Outside charge completed: " +
+                    "Failed charge number reset for member=${counter.memberId} to count=${counter.numberFailedCharges}"
+        }
     }
 
+    @ResetHandler
+    fun reset() = this.failedChargeCounterManager.dangerouslyDeleteAll()
```
meletios: btw, reseting this event handler, that means that we will send to Segment agian all the values. Is that ok? or are we going to end up with duplicates?

vonElfvin: That is what we want, good catch though!

PR #34, opened 2019-10-04
Add thread sleep to segment call with single threaded event handling

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/segment/SegmentManagerImpl.kt
```diff
@@ -20,5 +20,6 @@ class SegmentManagerImpl @Autowired constructor(
                 .enableIntegration("Customer.io", true)
                 .traits(traits)
         )
+        Thread.sleep(10)
```
meletios: Would be perfect to be a configuration this one.

PR #41, opened 2019-10-22
Backfill account entries

akashpatl:


src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
-    now: Instant
-  ) {
-    currentlyCheckingPeriod = period
+            productSubscriptionsOfPeriod.stream().forEach({o -> logger.info { "${o.productId} -- ${o.subscription}" }})
+            scheduleProductSubscriptions(memberId, period, eventType, productSubscriptionsOfPeriod, now)
+            scheduleCampaignDiscounts(memberId, period, eventType, discountOfPeriod, now)
+
+            //Don't schedule subscriptions for backfills
+            if (eventType is ScheduleSubscriptionEvent) {
+                addSubscriptionScheduled(eventType.memberId, period, eventType.scheduledBy, now)
```
vonElfvin: We should do this regardless actually. Since we should always schedule once per period.

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
+    private fun scheduleProductSubscriptions(
+        memberId: String,
+        period: YearMonth,
+        event: Any,
+        productSubscriptionsOfPeriod: List<InsuranceProductSubscriptionDto>,
+        now: Instant
+    ) {
+        if (productSubscriptionsOfPeriod.isNotEmpty()) {
+            productSubscriptionsOfPeriod.forEach { productSubscription ->
+                when (event) {
```
vonElfvin: We don't need to handle the events separately when scheduling subscriptions.

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
+        memberId: String,
+        period: YearMonth,
+        event: Any,
+        discountsOfPeriod: List<CampaignDiscount>,
+        now: Instant
+    ) {
+        if (discountsOfPeriod.isNotEmpty()) {
+            discountsOfPeriod.stream().forEach { discountOfPeriod ->
+                if (discountOfPeriod.amount.number.doubleValueExact() > 0L) {
+                    when (event) {
```
vonElfvin: Same goes for discounts

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
+                                "DISCOUNT entry added to memberId=${event.memberId} " +
+                                        "with reference=${discountOfPeriod.reference} " +
+                                        "an type={${discountOfPeriod.type.name}} " +
+                                        "in ${period.month} ${period.year} " +
+                                        "with amount=${discountOfPeriod.amount}"
+                            }
+
+                        }
+                        is ScheduleSubscriptionBackfillEvent -> {
+                            if (!accountEntryAlreadyExists(memberId, discountOfPeriod.reference.toString(), period, EntryType.fromCampaignDiscount(discountOfPeriod))) {
```
vonElfvin: This check should be way before with "SubscriptionScheduled". Since we never want to check a month more than once "SubscriptionScheduled" keeps track if we have checked that month before. Thus we should use the same logic both for regular scheduling and backfilling.

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
 
-  private fun shouldCheckPeriod(period: YearMonth, memberId: String, checkAllPeriods: Boolean = false): Boolean {
-    if (period.isBefore(firstPeriod)) {
-      return false
+    private fun accountEntryAlreadyExists(memberId: String, reference: String, period: YearMonth, type: EntryType): Boolean {
+        return accountManager.checkEntryExistsForPeriod(memberId, reference, period.atDay(1), type)
     }
-    if (!subscriptionAlreadyScheduled(memberId, period)) {
-      return true
+
```
vonElfvin: This is redundant since the method below is even more powerful (given we should never check the same period twice)

vonElfvin: Having a method checking the entries directly opens up for potential misscalculations of the subscriptions if someone for example moves after we have charged them then we have anotehr productId and thus we would charge them more with this solution.

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
+    private fun shouldCheckPeriod(
+        period: YearMonth,
+        memberId: String,
+        checkAllPeriods: Boolean = false,
+        event: Any
+    ): Boolean {
+        if (event is ScheduleSubscriptionEvent && period.isBefore(firstPeriod)) {
+            return false
+        }
+        if (event is ScheduleSubscriptionBackfillEvent && period.isBefore(firstPeriodBackfill)) {
```
vonElfvin: Even though I think this is really clever, sending "firstPeriod" instead of "event" and making one check is probably better since we should not treat the different events differently with the exception of what the first period is.

src/test/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSagaTest.kt
```diff
@@ -81,7 +81,8 @@ class SubscriptionScheduledSagaTest {
     fixture = SagaTestFixture(SubscriptionScheduledSaga::class.java)
     fixture.registerResource(productPricingService)
     fixture.registerResource(subscriptionScheduledManager)
-    val scheduleSubscriptionEvent = ScheduleSubscriptionEvent(MEMBER_ID, SCHEDULED_BY_MAIL, STATE)
+    val scheduleSubscriptionEvent =
+      ScheduleSubscriptionEvent(MEMBER_ID, SCHEDULED_BY_MAIL, STATE)
```
vonElfvin: Some more tests may be good. There are some cases that will fail with the current implementation (see accountEntryAlreadyExists comment)

src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -108,6 +108,21 @@ class ScheduleController @Autowired constructor(
         return ResponseEntity.noContent().build()
     }
 
+    @PostMapping("/backfill/{memberId}")
```
vonElfvin:
```suggestion
    @PostMapping("/backfill/subscription/{memberId}")
```

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
vonElfvin: Actually, this is really dangerous because now we will check all periods a second time when we schedule subscriptions regularly before charges 😅

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/sagas/SubscriptionScheduledSaga.kt
```diff
@@ -24,190 +27,295 @@ import java.time.Instant
...
+            checkSubscriptionsOfPeriod(eventType, previousPeriod, memberId, checkAllPeriods, now)
+        }
+    }
 
-    if (!subscriptionAlreadyScheduled(memberId, period)) {
-      val productSubscriptionsOfPeriod = getProductSubscriptions(memberId, period)
+    private fun scheduleProductSubscriptions(
+        memberId: String,
+        period: YearMonth,
+        event: Any,
```
vonElfvin: Something like this would probably be better since we don't want to have different logic for the different event types.

```suggestion
        stopAtPeriod: YearMonth,
```

PR #42, opened 2019-10-24
Add ability to dangerously delete entries

vonElfvin:
* Deletes account entries via an endpoint, to be used in rare occasions where something is obviously incorrect.


vonElfvin:
> Any particular reason we can't use a correction/reversal? :)

Yes, because in this case we have an entry of negative 9k that should never have showed up here in the first place. This should almost never be used, it's just for race occurances like these.

PR #43, opened 2019-10-28
Add ability to add campaign discount through endpoint

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -113,4 +116,19 @@ class MemberAccount() {
             }
         }
     }
+
+    @CommandHandler
+    fun handle(command: AddCampaignEntryCommand) {
+        val balance = getCurrentBalance()
+        if (balance.isNegativeOrZero) {
+            throw Exception("Cannot add a campaign entry to member=${command.memberId} because balance is $balance")
```
akashpatl: Pretty obvious, but maybe add that it can't be added because it is negative or zero.

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -113,4 +116,19 @@ class MemberAccount() {
...
+        }
+        AggregateLifecycle.apply(EntryAddedEvent(command.memberId, Entry.from(command, balance)))
```
akashpatl: This makes it seem like you're taking the sum of all account entries and giving them a discount equal to the negative of the total balance, i.e resetting their entire balance?

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -113,4 +116,19 @@ class MemberAccount() {
...
+    }
+
+    private fun getCurrentBalance(): MonetaryAmount {
+        return entries
+            .filter { entry -> entry.failedAt == null }
```
akashpatl: What entries are we including here? Don't we need to include a filter for the member Id?

src/test/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccountTest.kt
```diff
@@ -254,4 +251,86 @@ class MemberAccountTest {
             .expectNoEvents()
     }
 
+    @Test
```
akashpatl: Nice test! Should've started my review here. Clarifies my questions above! :D

PR #46, opened 2019-10-30
WIP guard against backfilling all completed charges from payment-service

vonElfvin:
* Does not reset the failed charges projection if the charge is in the past
* Does not apply an event if the completed charge has already been handled by the SchedulerAggregate
☝️ Already does the same in the MemberAccount aggregate


vonElfvin:
Out of date implementation, solved using something simpler

PR #50, opened 2019-11-19
Add ability to forgive debt for paying members that have debt from 2018

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -163,9 +165,44 @@ class MemberAccount() {
...
+        apply(EntryAddedEvent(memberId, entry))
+    }
+
     private fun getCurrentBalance(): MonetaryAmount {
         return entries
             .filter { entry -> entry.failedAt == null }
             .fold(Money.of(BigDecimal.ZERO, "SEK"), { sum, entry -> sum.add(entry.amount) })
     }
+
+    private fun getBalanceOfYear(year: Year): MonetaryAmount {
```
meletios: Shall we pass a localdate here in order to dont have the logic on line 205?

vonElfvin: I changed the logic of how the balance is returned for clarity, please check it out

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -22,7 +22,7 @@ data class Entry(
     val source: String,
     val reference: String,
     val addedBy: String = "system",
-    val addedAt: Instant,
+    val addedAt: Instant = Instant.now(),
```
palmenhq: are we sure about this? considering its event sourced `now` might be different if we replay the events

vonElfvin: Since it's not nullable it will always be present in the events and thus never use Instant.now() in replay. Or am I missing something?=

palmenhq: Then why do we need a default value? or what am i missing?

johantjellden: Yes I would agree with @palmenhq having default values on event classes could lead to really hard to detect bugs.

vonElfvin: I cannot see a scenario where the default value can be used when handling the event since addedAt is not nullable. But I changed it now regardless.

PR #52, opened 2019-11-27
Add endpoint to forgive debt that is more than 6 months old according to the law

vonElfvin:
* Instead of only forgiving 2018 debt of paying members we have to, by law, forgive all debt that is more than 6 months old.
* Adds debt older than 6 months ago as realized LOSS so we don't charge for money we aren't allowed to charge


src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -168,19 +168,16 @@ class MemberAccount() {
     }
 
     @CommandHandler
-    fun handle(command: MaybeAddLossEntryToForgive2018BalanceCommand) {
-        val balance2018 = getBalanceOfYear(Year.of(2018))
+    fun handle(command: MaybeAddLossEntryToForgiveOldDebtCommand) {
+        val periodBeforeHalfYearAgo = YearMonth.now().minusMonths(7)
```
meletios: Half year or 7 months? 🤷‍♂️

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -168,19 +168,16 @@ class MemberAccount() {
...
+        val balanceUntilHalfYearAgo = getBalanceOPeriod(periodBeforeHalfYearAgo)
         val currentBalance = getCurrentBalance()
-        if (command.numberFailedCharges != 0) {
-            logger.info { "Did NOT forgive debt of member=$memberId since it does not have 0 failed charges (numberFailedCharges=${command.numberFailedCharges})" }
+        if (balanceUntilHalfYearAgo.isNegativeOrZero) {
+            logger.info { "Did NOT forgive debt of member=$memberId since it has a 2018 balance that is negative or zero (2018 balance=${balanceUntilHalfYearAgo})" }
```
meletios: Is this error message true? Since YearMonth.now().minusMonths(7) is March 2019

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -168,19 +168,16 @@ class MemberAccount() {
...
     @CommandHandler
-    fun handle(command: MaybeAddLossEntryToForgive2018BalanceCommand) {
-        val balance2018 = getBalanceOfYear(Year.of(2018))
+    fun handle(command: MaybeAddLossEntryToForgiveOldDebtCommand) {
+        val periodBeforeHalfYearAgo = YearMonth.now().minusMonths(7)
+        val balanceUntilHalfYearAgo = getBalanceOPeriod(periodBeforeHalfYearAgo)
         val currentBalance = getCurrentBalance()
-        if (command.numberFailedCharges != 0) {
-            logger.info { "Did NOT forgive debt of member=$memberId since it does not have 0 failed charges (numberFailedCharges=${command.numberFailedCharges})" }
+        if (balanceUntilHalfYearAgo.isNegativeOrZero) {
```
akashpatl: a bit of a confusing name. Do you simply mean "balance half a year ago"?

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
akashpatl: I assume this is a job we'll have to run periodically since we forgive everything Now - 6 months as a rolling window. Does it make sense to write 2018? Maybe calculate the period instead?

vonElfvin: It is not! Nice catch

src/main/kotlin/com/hedvig/accountservice/accounts/web/LossController.kt
```diff
@@ -17,18 +14,31 @@ class LossController @Autowired constructor(
     private val failedChargeCounterManager: FailedChargeCounterManager,
     private val commandGateway: CommandGateway
 ) {
-    @PostMapping("/forgive/2018/payers")
-    fun forgive2018DebtForPeopleWhoPay(
+    @PostMapping("/forgive/outdated/debt")
+    fun forgiveOldDebt(
```
akashpatl: Small consistency check. Is there a difference between old and outdated debt? I personally prefer outdated.

src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
vonElfvin: everything up until 7 months ago. The month 6 months ago is half a year ago and we want the balance the month before that.

src/test/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccountTest.kt
```diff
@@ -623,118 +627,76 @@ class MemberAccountTest {
...
             id = UUID.randomUUID(),
-            fromDate = LocalDate.of(2018, 11, 1),
+            fromDate = today.minusMonths(8).withDayOfMonth(1),
             amount = Money.of(BigDecimal.ONE, "SEK"),
-            type = EntryType.CHARGE,
+            type = EntryType.SUBSCRIPTION,
             source = "product-pricing",
             reference = "productId",
             title = "Monthly fee",
             addedAt = Instant.now()
```
akashpatl: Shouldn't addedAt be a timestamp 8 months ago? It might not affect the test whatsoever now but if we ever introduce new logic that depends on that field, our tests should break/or pass according to what case the test is trying to simulate in real life. Just an idea :)

PR #53, opened 2019-11-27
Dangerously reset scheduler

vonElfvin:
Add endpoint giving ability to dangerously reset scheduler and delete subscriptions of member

To be used to restart process of members that we know are not started properly. Should never be used unless we have to.


src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -193,6 +193,26 @@ class MemberAccount() {
         logger.info { "Loss entry ADDED to member=$memberId to forgive outdated debt as of $outdatedPeriod=$balanceOutdatedPeriod" }
     }
 
+    @CommandHandler
+    fun handle(command: DangerouslyDeleteAllSubscriptionEntriesCommand) {
+        entries
+            .filter { entry ->
+                entry.type == EntryType.SUBSCRIPTION ||
```
meletios: Should we delete `CAMPAIGN`, `CORRECTION` or `CHARGE` also?

vonElfvin: We only want to delete subscription, free months and referrals. These are the ones fetched by account-service and the only ones to be removed here. 

Everything else is either manual or a charge and should not be removed.

PR #54, opened 2019-11-28
Return 400 instead of crashing when invalid reset is sent

vonElfvin:


src/main/kotlin/com/hedvig/accountservice/schedulers/web/ScheduleController.kt
```diff
@@ -309,9 +310,13 @@ class ScheduleController @Autowired constructor(
...
         @PathVariable memberId: String,
         @RequestParam resetBy: String
-    ) {
-        commandGateway.sendAndWait<Void>(
+    ): ResponseEntity<SchedulerStateTransitionDto> {
+        val transition = commandGateway.sendAndWait<SchedulerStateTransition>(
             DangerouslyResetCommand(memberId, resetBy)
         )
+        if (transition.outcome != SchedulerStateTransitionOutcome.ALLOWED_TRANSITION) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(SchedulerStateTransitionDto.from(transition))
```
meletios: 405 instead of 400? Minor but 😁

vonElfvin: 👍

PR #56, opened 2019-11-28
Fail more gracefully when exception is thrown when maybe terminating member

vonElfvin:
While also fixing a reason for it to fail 👼


src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -41,13 +44,10 @@ class NumberFailedChargesProjection @Autowired constructor(
...
         }
         if (!replayStatus.isReplay) {
-            val terminationOutcome = terminationService.maybeTerminateMember(event.memberId)
-            if (terminationOutcome == TerminationOutcome.NOT_QUALIFIED) {
-                this.notificationService.sendFailedChargeInsufficientFundsNotification(event.memberId)
-            }
-            if (terminationOutcome == TerminationOutcome.TERMINATED) {
-                this.notificationService.sendFailedChargesTerminationNotification(event.memberId)
-            }
+            notifyAndMaybeTerminateMember(
```
meletios: Bra jobbt

PR #68, opened 2020-04-01
Feature/estimation norway

meletios:


src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -12,110 +12,119 @@ import org.springframework.beans.factory.annotation.Autowired
...
+    )
+  }
+
+  private fun getEstimatedSubscription(memberId: String, period: YearMonth): MonetaryAmount {
+    val currentProductSubscriptions =
+      productPricingService.getMonthlyPremiumPerPeriod(memberId, period)
+    if (currentProductSubscriptions.isEmpty()) {
+      return Money.of(0, "SEK") //TODO: We dont know the country
     }
+    val currency = currentProductSubscriptions.first().cost.currency
```
FredrikAreschoug, line 83: Could also do a check her if all currencys is equal to first currency

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -12,110 +14,130 @@ import org.springframework.beans.factory.annotation.Autowired
...
+      subscription = estimatedSubscription,
+      discount = estimatedDiscount,
+      charge = estimatedCharge,
+      discountCodes = discountCodes
+    )
+  }
+
+  private fun getEstimatedSubscription(memberId: String, period: YearMonth): MonetaryAmount {
+    val currentProductSubscriptions =
+      productPricingService.getMonthlyPremiumPerPeriod(memberId, period)
```
vonElfvin, line 82: getMonthlyPremiumCostPerPeriod

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -12,110 +14,130 @@ import org.springframework.beans.factory.annotation.Autowired
...
+    if (currentProductSubscriptions.isEmpty()) {
+      return Money.of(0, "SEK") //TODO: We dont know the country
     }
```
vonElfvin, line 85: There is an endpoint on product-pricing that returns a "ContractMarketInfo" with two properties: market and preferredCurrency. So use that endpoint with the memberId and you'll get a currenc

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -12,110 +14,130 @@ import org.springframework.beans.factory.annotation.Autowired
...
-    private fun getEstimatedDiscount(memberId: String, period: YearMonth): MonetaryAmount {
-        val currentDiscounts = productPricingService.getMonthlyCampaignDiscounts(memberId, period)
-        return currentDiscounts
-            .map { discount -> discount.amount }
-            .fold(Money.of(0, "SEK")) { total, discount -> total.add(discount) }
+  private fun getEstimatedDiscount(memberId: String, period: YearMonth): MonetaryAmount {
+    val currentDiscounts = productPricingService.getMonthlyCampaignDiscounts(memberId, period)
+    if (currentDiscounts.isEmpty()) {
+      return Money.of(0, "SEK") //TODO: We dont know the country
     }
```
vonElfvin, line 96: use currency from above

vonElfvin: I.e. if there is a subscription you know the currency, or use the endpoint

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -12,110 +14,130 @@ import org.springframework.beans.factory.annotation.Autowired
...
+      currentMonth.plusMonths(1)
+    }
+  }
+
+  private fun getCurrency(
+    memberId: String,
+    listOfAmounts: List<MonetaryAmount>
+  ): CurrencyUnit {
+    val distinctCurrencies = listOfAmounts
+      .map { o -> o.currency }.toSet()
```
vonElfvin, line 139: There is something called "distinct()" as well in kotlint instead of toSet, reads a bit more nicely IMO

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingClient.kt
```diff
@@ -39,4 +49,11 @@ interface ProductPricingClient {
     @PathVariable memberId: String,
     @RequestBody request: SetCancellationDateRequest
   )
+
+  @GetMapping("/_/premiums/cost/member/{memberId}/period/{period}")
+  fun getMonthlyPremiumPerPeriod(
```
vonElfvin, line 54: getMonthlyPremiumCostForPeriod

vonElfvin: In my head "premium" is what you pay "per month" where as premium cost is what you actually should pay for the given period.

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/ProductPricingServiceImpl.kt
```diff
@@ -35,9 +40,15 @@ class ProductPricingServiceImpl @Autowired constructor(
...
   override fun cancelProduct(memberId: String, productId: UUID, terminationDate: LocalDate) {
-    productPricingClient.setCancellationDate(memberId, SetCancellationDateRequest(
-      insuranceId = productId,
-      inactivationDate = terminationDate.atStartOfDay().atZone(ZoneId.of("Europe/Stockholm")).toInstant()
-    ))
+    productPricingClient.setCancellationDate(
+      memberId, SetCancellationDateRequest(
+        insuranceId = productId,
+        inactivationDate = terminationDate.atStartOfDay().atZone(ZoneId.of("Europe/Stockholm")).toInstant()
+      )
```
vonElfvin, line 47: In the future we should change this as well to terminate using contract endpoint. Not something you should do now though

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/dto/AgreementPremiumCost.kt
```diff
@@ -0,0 +1,9 @@
+package com.hedvig.accountservice.integration.productPricing.dto
+
+import java.util.UUID
+import javax.money.MonetaryAmount
+
+class AgreementPremiumCost(
+  val agreementId: UUID,
+  val cost: MonetaryAmount
+)
```
vonElfvin, line 9: We should add "lineOfBusiness" to this DTO, I think Fredrik has already done it but otherwise we really should do it. Both in product-pricing and account-service.

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
vonElfvin: In my head "premium" is what you pay "per month" where as premium cost is what you actually should pay for the given period.

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
meletios: sweet, i missed that

PR #69, opened 2020-04-01
Feature/estimation norway

meletios:


src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountManager.kt
```diff
@@ -58,14 +59,18 @@ class AccountManager @Autowired constructor(
   fun getNextChargeEstimation(memberId: String): ChargeEstimationDto {
     val monthToCheckMaybe = getMonthForEstimation(memberId)
     val currentBalance = getCurrentBalance(memberId)
+    val marketInfo = productPricingService.getMarketInfoForMember(memberId)
```
vonElfvin, line 62: To make the methods more easily readable
```suggestion
    val currency = productPricingService.getMarketInfoForMember(memberId).preferredCurrency
```

meletios:
> I'd say "preferredCurrency" should be used when the list is empty, and the hirarachy should be like this:
> 
>     * If there is currency available for both subscription and discount, use their currencies
> 
>     * If there is a subscription list available but no discount, use that `currency` for discount, i.e. `discount = Money.of(0, subscriptionCurrency)`
> 
>     * If there is no subscription and no discount, i.e. `subscription = Money.of(0, preferredCurrency)` and `discount = Money.of(0, preferredCurrency)`

Dont agree with that, since i am checking if the "preferredCurrency" is the same with all other currencies

PR #71, opened 2020-04-16
added bundle discount

FredrikAreschoug:


src/main/kotlin/com/hedvig/accountservice/accounts/model/EntryType.kt
```diff
@@ -12,6 +12,7 @@ enum class EntryType {
   REFERRAL_DISCOUNT,
   FREE_MONTH_DISCOUNT,
   PERCENTAGE_MONTH_DISCOUNT,
+  BUNDLE_DISCOUNT,
```
FredrikAreschoug, line 15: What is this used for? should we have `BUNDLE_DISCOUNT_COST_DEDUCTION` and `BUNDLE_DISCOUNT_PERCENTAGE_DEDUCTION`

vonElfvin: I think that sounds wise actually, since we already know we will have 2. These are used to trace what kind of discount is applied.

PR #72, opened 2020-04-22
Store currency

FredrikAreschoug:


src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt
```diff
@@ -268,7 +269,9 @@ class MemberAccount() {
         return entries
             .filter { entry -> entry.type == EntryType.LOSS }
             .filter { entry -> entry.fromDate.isAfter(period.atEndOfMonth()) }
-            .fold(Money.of(BigDecimal.ZERO, "SEK"), { sum, entry -> sum.add(entry.amount)})
+            .map { it.amount }
+            .ifEmpty { listOf(Money.of(BigDecimal.ZERO, "SEK")) }
```
vonElfvin, line 273:
```suggestion
            .ifEmpty { listOf(Money.of(BigDecimal.ZERO, getCurrency())) }
```

And then do something like this
```
fun getCurrency() = entries.map { entry -> entry.amount.currency }.distinct().first()
``
`

File: src/main/kotlin/com/hedvig/accountservice/accounts/domain/MemberAccount.kt [hunk 2]
```diff
@@ -284,10 +287,14 @@ class MemberAccount() {
...
+        .map { it.amount }
+        .ifEmpty { listOf(Money.of(BigDecimal.ZERO, "SEK")) }
+        .reduce(MonetaryAmount::add)
 
     private fun hasAnyCompletedChargeAfter(period: YearMonth): Boolean {
         val dateAtEndOfPeriod = period.atEndOfMonth()
         return entries.any { entry -> entry.type == EntryType.CHARGE && entry.chargedAt != null && entry.fromDate.isAfter(dateAtEndOfPeriod) }
     }
+
+
```
vonElfvin, line 299: ktlint

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountBalance.kt
```diff
@@ -1,8 +1,8 @@
 package com.hedvig.accountservice.accounts.model
 
-import javax.money.MonetaryAmount
+import java.math.BigDecimal
 
 data class AccountBalance(
     val memberId: String,
-    val currentBalance: MonetaryAmount
-)
+    val currentBalance: BigDecimal
```
vonElfvin, line 7: What about the currency?

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -17,7 +22,9 @@ data class Entry(
     val id: UUID = UUID.randomUUID(),
     val memberId: String,
     val fromDate: LocalDate,
-    val amount: MonetaryAmount,
+    @ColumnName("amount")
```
vonElfvin, line 25: Smart!

src/main/kotlin/com/hedvig/accountservice/accounts/model/Entry.kt
```diff
@@ -29,6 +36,42 @@ data class Entry(
     var chargedAt: Instant? = null,
     var failedAt: Instant? = null
 ) {
+    val amount: MonetaryAmount
+        get() = Money.of(amountBigDecimal, amountCurrency)
```
vonElfvin, line 40: Really neat

src/main/kotlin/com/hedvig/accountservice/extensions/MonetaryAmountExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.hedvig.accountservice.extensions
+
+import java.math.BigDecimal
+import javax.money.MonetaryAmount
+
+fun MonetaryAmount.asBigDecimal(): BigDecimal {
+    return this.number.numberValue(BigDecimal::class.java)
+}
```
vonElfvin, line 8: This one should be in some common util library, I create this extension everywhere as well

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SchedulerState.kt
```diff
@@ -10,6 +13,27 @@ data class SchedulerState(
     val status: ChargeStatus,
     val changedBy: String,
     val changedAt: Instant = Instant.now(),
-    val amount: MonetaryAmount? = null,
+    val amountBigDecimal: BigDecimal? = null,
```
vonElfvin, line 16: Do we need
```suggestion
     @ColumnName("amount") 
    val amountBigDecimal: BigDecimal? = null,
```
Here?

FredrikAreschoug: It's stored in `SchedulerStateRepository` so I think we will need to

src/main/kotlin/com/hedvig/accountservice/accounts/model/AccountBalance.kt
FredrikAreschoug: Got a lot of problems trying to add it to the query for `sumAllBalancesUntilDateExcludingFailedCharges`. So then I Investigates where that function where used an it's only in test so I figured let just go with using BigDecimal. What do you think should I try to fix it? what is it used for?

vonElfvin: Ah okay, no don't bother fixing that

src/main/kotlin/com/hedvig/accountservice/schedulers/model/SchedulerState.kt
vonElfvin: I was referring to the @ColumnName

FredrikAreschoug: I think so otherwise, jdbi will look for the column `amount_big_decimal` to instantiate this object :)

PR #75, opened 2020-04-27
revert blacklist

meletios:


src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -1,8 +1,50 @@
 package com.hedvig.accountservice.schedulers.domain
 
 import com.hedvig.accountservice.extensions.asBigDecimal
-import com.hedvig.accountservice.schedulers.domain.commands.*
-import com.hedvig.accountservice.schedulers.domain.events.*
+import com.hedvig.accountservice.schedulers.domain.commands.ApproveChargeCommand
```
FredrikAreschoug, line 4: Whow 1,018 guessing all of them are correct :smile:

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -12,485 +54,509 @@ import org.axonframework.commandhandling.CommandHandler
...
+  }
+
+  @CommandHandler
+  fun handle(command: RevertBlacklistSchedulerCommand): SchedulerStateTransition {
+    val nextState = SchedulerState(
+      memberId = command.memberId,
+      status = ChargeStatus.INITIATED,
+      changedBy = command.revertedBy
+    )
+    val transition = isAllowedToTransition(nextState)
```
vonElfvin, line 524: This will probably break since you are not allowed to move from blacklist to any other state. I think we only need a simple check as 
```suggestion
    if (currentState != BlackListed) throw
```

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -12,485 +54,509 @@ import org.axonframework.commandhandling.CommandHandler
...
+    if (transition.outcome != SchedulerStateTransitionOutcome.ALLOWED_TRANSITION) {
+      apply(BlacklistSchedulerDeniedEvent(command.memberId, currentState, nextState))
+      logger.error { "Reverting Blacklist is DENIED for memberId=${command.memberId} (Message: ${transition.message})" }
+      return transition
+    }
```
vonElfvin, line 529: I would just remove this with the check above instead

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -12,485 +54,509 @@ import org.axonframework.commandhandling.CommandHandler
...
+    apply(SchedulerInitiatedAfterBlacklistRevertedEvent(command.memberId, nextState))
```
vonElfvin, line 530: Not sure about the naming, since the event always contains a "nextState" property I think it's sufficient with 
```suggestion
    apply(SchedulerBlacklistRevertedEvent(command.memberId, nextState))
```

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -12,485 +54,507 @@ import org.axonframework.commandhandling.CommandHandler
...
+
+  @CommandHandler
+  fun handle(command: RevertBlacklistSchedulerCommand): SchedulerStateTransition {
+    val nextState = SchedulerState(
+      memberId = command.memberId,
+      status = ChargeStatus.INITIATED,
+      changedBy = command.revertedBy
+    )
+    require(currentState.status == ChargeStatus.BLACKLISTED) {
+      "Tried to revert a Blacklisted state but the status is ${currentState.status} [MemberId: ${currentState.memberId}]"
```
vonElfvin, line 525: I would rather just return a SchedulerStateTransitionOutcome instead of throwing here

vonElfvin: Since we have "transition.outcome != SchedulerStateTransitionOutcome.ALLOWED_TRANSITION" outside

meletios: Check again 😁

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -12,485 +54,508 @@ import org.axonframework.commandhandling.CommandHandler
...
+
+  @CommandHandler
+  fun handle(command: RevertBlacklistSchedulerCommand): SchedulerStateTransition {
+    val nextState = SchedulerState(
+      memberId = command.memberId,
+      status = ChargeStatus.INITIATED,
+      changedBy = command.revertedBy
+    )
+    if (currentState.status != ChargeStatus.BLACKLISTED) {
+      logger.info("Tried to revert a Blacklisted state but the status is ${currentState.status} [MemberId: ${currentState.memberId}]")
```
vonElfvin, line 525:
```suggestion
      logger.error("Tried to revert a Blacklisted state but the status is ${currentState.status} [MemberId: ${currentState.memberId}]")
```

PR #82, opened 2020-06-17
Feature/read failed charges from event store

FredrikAreschoug:
Fix for termination bug:
* replace failed charge counter logic (incrementing failed charges by 1 on each failed charge) by reading number of continuous failed charges from events
* Get contracts instead of product from Product-pricing using getContractsByMemberId and terminateContract rather than cancelProduct
* call failedChargeCalculator from getNumberOfFailedCharges endpoint (rather than just reading from db) so that we can trigger this calculation from somewhere other than in the payment flow, so that we can check that we are getting the correct numberFailedCharges 

Termination logic:
* if contract has Market.Sweden -> we will terminate if a member has 3 or more failed payments and a flag or 6 or more failed payments without a flag.
* if contract has Market.Norway -> we will terminate if member has 6 or more failed payments - (we do not have the concept of flags on a member for Norway)

Please let us know your thoughts :)


src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+        return listDomainEventMessages.takeLastWhile {
+            filterFailedEventTerminateMember.contains(it.type)
+        }.size
+    }
+
+    private val filterFailedEventTerminateMember = listOf(
+        ChargeFailedEvent::class.java.simpleName,
+        ChargeRequestFailedEvent::class.java.simpleName
+    )
+    private val filterListTerminateMember = mutableListOf(
```
vonElfvin, line 24: What's with the name of this? Why "terminateMember"

src/test/kotlin/com/hedvig/accountservice/accounts/util/EventStoreUtilsTest.kt
```diff
@@ -0,0 +1,292 @@
...
+            a.ChargeRequestFailedEventBuilder(memberId, today)
+        )
+
+        val result = EventStoreUtils.numberOfLastContinuousFailedCharges(memberId, eventStore)
+
+        Assertions.assertThat(result).isEqualTo(4)
+    }
+
+    @Test
+    fun `when last 4 charge events are the order -ChargeEntryFailedEvent, ChargeEntryFailedEvent, ChargeCompletedEvent and ChargeEntryFailedEvent- numberOfLastContinuousFailedCharges should return 1`() {
```
vonElfvin, line 75: ChargeEntryFailedEvent is unrelated to the count of number failed charges

src/test/kotlin/com/hedvig/accountservice/accounts/util/EventStoreUtilsTest.kt
```diff
@@ -0,0 +1,292 @@
...
+import io.mockk.junit5.MockKExtension
+import org.assertj.core.api.Assertions
+import org.axonframework.eventsourcing.eventstore.EventStore
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import java.sql.Timestamp
+import java.time.*
+
+@ExtendWith(MockKExtension::class)
+class EventStoreUtilsTest {
```
vonElfvin, line 16: Maybe add some random event just to see that it correctly filters them out, for example ChargeInitiatedEvent

src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+        }.asStream().toList().sortedBy {
+            it.timestamp
+        }
+
+        return listDomainEventMessages.takeLastWhile {
+            filterFailedEventTerminateMember.contains(it.type)
+        }.size
+    }
+
+    private val filterFailedEventTerminateMember = listOf(
```
vonElfvin, line 20: These names should probably be clarified, terminating members is not the core of this counter, the count itself is the core and it's used as input for a potential termination.

src/test/kotlin/com/hedvig/accountservice/accounts/util/EventStoreUtilsTest.kt
FredrikAreschoug: Yeah I and @AlexSJ1990 did a mix up when we wrote the test list, we should change it!

src/test/kotlin/com/hedvig/accountservice/accounts/util/EventStoreUtilsTest.kt
FredrikAreschoug: but we did, not in all places but `createEventThatShouldBeFilteredOut` creates `ChargeInitiatedEvent`

src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
FredrikAreschoug: Yes, def! We were going with the logic of numberOfLastContinuousFailedCharges returning a bool in the beginning but forgot to rename this :slightly_smiling_face:

src/test/kotlin/com/hedvig/accountservice/accounts/util/EventStoreUtilsTest.kt
AlexSJ1990: we have added a few ChargeInitiatedEvents in some of the tests already 🙂

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/FailedChargeCounterManager.kt
```diff
@@ -22,19 +22,11 @@ class FailedChargeCounterManager @Autowired constructor(
             repository.getFailedChargeCounter(memberId)
         }
 
-    fun incrementNumberFailedCharges(memberId: String, timestamp: Instant): FailedChargeCounter =
+    fun updateNumberFailedCharges(memberId: String, numberFailedCharges: Int, timestamp: Instant): FailedChargeCounter =
         jdbi.withHandle<FailedChargeCounter, RuntimeException> { h ->
```
johantjellden, line 26: There is a `withHandleUnchecked` extension method that is pretty nice. With that you don't have to specify the types :D

src/test/kotlin/com/hedvig/accountservice/builder/a.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.hedvig.accountservice.schedulers.model.SchedulerState
+import org.axonframework.eventsourcing.GenericDomainEventMessage
+import org.axonframework.eventsourcing.eventstore.DomainEventStream
+import org.javamoney.moneta.Money
+import java.time.Instant
+import java.util.*
+import javax.money.MonetaryAmount
+
+object a {
+    fun domainEventStream(vararg eventBuilders: EventBuilder) =
```
johantjellden, line 14: 💯

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -50,13 +50,12 @@ class NumberFailedChargesProjection @Autowired constructor(
         replayStatus: ReplayStatus,
         domainEventMessage: DomainEventMessage<Any>
     ) {
-        if (timestamp.isBefore(failedChargeCounterManager.getFailedChargeCounter(event.memberId)!!.lastUpdateAt)) {
+        val counter = failedChargeCalculator.calculateAndUpdateNumberOfFailedCharges(event.memberId, timestamp)
+
+        if (timestamp.isBefore(counter.lastUpdateAt)) {
```
AlexSJ1990, line 55: is this needed anymore?

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -126,20 +117,25 @@ class NumberFailedChargesProjection @Autowired constructor(
...
+        val contracts = productPricingService.getContractsByMemberId(memberId)
+
+        contracts.forEach { contract ->
+            try {
+                val terminationOutcome = terminationService.maybeTerminateContract(contract, memberId, timestamp)
+                if (terminationOutcome == TerminationOutcome.NOT_QUALIFIED) {
+                    notification(memberId, requestId)
+                }
+                if (terminationOutcome == TerminationOutcome.TERMINATED) {
+                    this.notificationService.sendFailedChargesTerminationNotification(memberId, requestId)
```
AlexSJ1990, line 129: what about if terminationOutcome is TerminationOutcome.NotActive ?

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -126,20 +117,25 @@ class NumberFailedChargesProjection @Autowired constructor(
...
+                }
+            } catch (exception: Exception) {
```
AlexSJ1990, line 131: why do we assume that any exception means that a member's direct debit has not been connected?

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -25,65 +24,113 @@ class TerminationService @Autowired constructor(
...
+                        logger.info {
+                            "Terminating contract ${contract.id} of member $memberId since " +
+                                    "the member has a Norwegian contract and $numberFailedCharges failed charges"
+                        }
+                        terminateProductOnLastDayOfMonth(
+                            memberId,
+                            timestamp,
+                            contract,
+                            numberFailedCharges,
+                            PersonFlags(Flag.NO_FLAG_PRESENT, Flag.NO_FLAG_PRESENT),
```
AlexSJ1990, line 49: is there a better way to do this?

src/main/kotlin/com/hedvig/accountservice/integration/productPricing/dto/ContractDto.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.util.*
+
+data class ContractDto(
+    val id: UUID,
+    @get:JsonProperty("isTerminated")
+    val isTerminated: Boolean,
+    val market: Market
+)
```
vonElfvin, line 11: Should probably check the masterInception before terminating

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
vonElfvin: Since you always update the counter now this would always return false. So I'd say remove it then.

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -75,9 +74,10 @@ class NumberFailedChargesProjection @Autowired constructor(
         replayStatus: ReplayStatus,
         domainEventMessage: DomainEventMessage<Any>
     ) {
-        val counter = failedChargeCounterManager.incrementNumberFailedCharges(event.memberId, timestamp)
-        logger.info {
-            "Charge request failed: Failed charge number incremented for member=${counter.memberId} to count=${counter.numberFailedCharges}"
```
vonElfvin, line 80: Please don't remove this logging. It's been incredibly useful for debugging and bug hunting.

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -75,9 +74,10 @@ class NumberFailedChargesProjection @Autowired constructor(
...
+        val counter = failedChargeCalculator.calculateAndUpdateNumberOfFailedCharges(event.memberId, timestamp)
+
+        if (timestamp.isBefore(counter.lastUpdateAt)) {
```
vonElfvin, line 79: Could remove this as well

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
FredrikAreschoug: isn't `calculateAndUpdateNumberOfFailedCharges` using `timestamp` to store `lastUpdateAt`? wich means they are the same?

FredrikAreschoug: so no it would be strange to keep it

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
vonElfvin: Then we shouldn't send any notifications

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/NumberFailedChargesProjection.kt
```diff
@@ -126,20 +117,25 @@ class NumberFailedChargesProjection @Autowired constructor(
...
-            }
-        } catch (exception: Exception) {
-            Sentry.capture(exception)
-            logger.error { "Error when checking if member should be terminated (Exception: $exception)" }
-            val hasActiveProduct = productPricingService.hasActiveProduct(memberId)
-            if (hasActiveProduct) {
-                this.notificationService.sendFailedChargeNotConnectedDirectDebitNotification(memberId, requestId)
+        val contracts = productPricingService.getContractsByMemberId(memberId)
+
+        contracts.forEach { contract ->
```
vonElfvin, line 122: Didn't you move this to a separate event listener?

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -25,65 +24,113 @@ class TerminationService @Autowired constructor(
...
+
+        return when (contract.market) {
+            Market.NORWAY -> {
+                when {
+                    numberFailedCharges >= reallyBadNumberOfFailedCharges -> {
+                        logger.info {
+                            "Terminating contract ${contract.id} of member $memberId since " +
+                                    "the member has a Norwegian contract and $numberFailedCharges failed charges"
+                        }
+                        terminateProductOnLastDayOfMonth(
```
vonElfvin, line 44:
```suggestion
                        terminateContractOnLastDayOfMonth(
```

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -25,65 +24,113 @@ class TerminationService @Autowired constructor(
...
         val lastDayOfMonth = accountManager.getLatestSubscriptionEntry(memberId)!!.fromDate
             .plusMonths(1)
             .withDayOfMonth(1)
             .minusDays(1)
-        productPricingService.cancelProduct(memberId, product.productId, lastDayOfMonth)
+
+        val terminationRequest = TerminateContractRequest(
+            timestamp.atZone(ZoneId.of("Europe/Stockholm")).toLocalDate(),
+            terminationReason,
+            null,
```
vonElfvin, line 127: A slight improvement here would be to include the number failed charges, the debt flag and the fraud flag upon termination as a comment. It's not perfect but at least we can see the reasoning.

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
FredrikAreschoug: I think this logic should be on `getPersonFlagsByMemberId(memberId)` instead, doing it here would mean that we would have to implement this in two places when adding it for norway

src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
```diff
@@ -0,0 +1,29 @@
+package com.hedvig.accountservice.utils
+
+import com.hedvig.accountservice.schedulers.domain.events.*
+import org.axonframework.eventsourcing.eventstore.EventStore
+import kotlin.streams.toList
+
+object EventStoreUtils {
+    fun numberOfLastContinuousFailedCharges(memberId: String, eventStore: EventStore): Int {
+        val listDomainEventMessages = eventStore.readEvents(memberId).filter { event ->
```
vonElfvin, line 9: Maybe the `filterMemberChargeEvents` could instead be a method that performs this filtering

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -25,65 +24,113 @@ class TerminationService @Autowired constructor(
...
-                terminateProductOnLastDayOfMonth(memberId, timestamp, currentProduct, numberFailedCharges, personFlags)
-                TerminationOutcome.TERMINATED
-            }
-            numberFailedCharges >= reallyBadNumberOfFailedCharges -> {
-                logger.info {
-                    "Terminating product ${currentProduct.productId} of member $memberId since " +
-                            "the member has $numberFailedCharges failed charges"
+
+        return when (contract.market) {
+            Market.NORWAY -> {
```
FredrikAreschoug, line 37: I do not think we should do this market separation, or what is the reason for it?

src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+            filterMemberChargeEvents.contains(event.type)
```
vonElfvin, line 10: And this line could also be a method instead  of using `filterMemberChargeFailedEvents` to instead check `isFailedCharge(event)`

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
AlexSJ1990: gooood point

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -69,21 +91,107 @@ class TerminationService @Autowired constructor(
...
+        terminationReason: TerminationReason
     ) {
         val lastDayOfMonth = accountManager.getLatestSubscriptionEntry(memberId)!!.fromDate
             .plusMonths(1)
             .withDayOfMonth(1)
             .minusDays(1)
-        productPricingService.cancelProduct(memberId, product.productId, lastDayOfMonth)
+
+        val terminationRequest = TerminateContractRequest(
+            timestamp.atZone(ZoneId.of("Europe/Stockholm")).toLocalDate(),
```
vonElfvin, line 185: We should terminate on `lastDayOfMonth` not today

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -69,21 +91,107 @@ class TerminationService @Autowired constructor(
...
+            terminationReason,
+            "member has $numberFailedCharges failed charges, " +
+                    "a debt flag of ${personFlags.debtFlag} and a fraud flag of ${personFlags.fraudFlag}",
+            false
+        )
+        productPricingService.terminateContract(contract.id, terminationRequest, null)
         customerIO.notifyMemberTerminated(
             userId = memberId,
-            productId = product.productId,
+            productId = contract.id,
```
vonElfvin, line 194:
```suggestion
            contractId = contract.id,
```

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
```diff
@@ -69,21 +91,107 @@ class TerminationService @Autowired constructor(
...
     }
 
-    private fun terminateProductOnLastDayOfMonth(
+    private fun notifyAndMaybeTerminateMember(
+        memberId: String,
+        requestId: String,
+        timestamp: Instant,
+        notification: (memberId: String, requestId: String) -> Unit
+    ) {
+        val contracts = productPricingService.getContractsByMemberId(memberId)
```
vonElfvin, line 100: Maybe filter out only active contracts for clarity?
```suggestion
        val contracts = productPricingService.getContractsByMemberId(memberId).filter(contract -> isActive(contract))
```

src/main/kotlin/com/hedvig/accountservice/utils/EventStoreUtils.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.hedvig.accountservice.schedulers.domain.events.*
+import org.axonframework.eventsourcing.DomainEventMessage
+import org.axonframework.eventsourcing.eventstore.DomainEventStream
+import org.axonframework.eventsourcing.eventstore.EventStore
+import kotlin.streams.toList
+
+object EventStoreUtils {
+    fun numberOfLastContinuousFailedCharges(memberId: String, eventStore: EventStore): Int {
+        val listDomainEventMessages = filterForChargeEvents(eventStore.readEvents(memberId))
```
vonElfvin, line 11:
```suggestion
        val chargeEventMessages = filterForChargeEvents(eventStore.readEvents(memberId))
```

src/main/kotlin/com/hedvig/accountservice/termination/TerminationService.kt
vonElfvin: The function is called `terminateContractOnLastDayOfMonth` for a reason ;)

vonElfvin: Adding names to call arguments would be a good improvement as well
```suggestion
           terminationDate = timestamp.atZone(ZoneId.of("Europe/Stockholm")).toLocalDate(),
```

src/main/kotlin/com/hedvig/accountservice/integration/memberService/dto/Flag.kt
```diff
@@ -3,5 +3,6 @@ package com.hedvig.accountservice.integration.memberService.dto
 enum class Flag {
     GREEN,
     AMBER,
-    RED
+    RED,
+    NO_FLAG_PRESENT
```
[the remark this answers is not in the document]
johantjellden: I think a problem here is that we are using member-service directly in our internal logic. Creating a class for our usecase and not allowing the member-service dto to escape out of this package might be another idea.

src/main/kotlin/com/hedvig/accountservice/projections/numberFailedCharges/FailedChargeCalculator.kt
```diff
@@ -0,0 +1,19 @@
...
+@Service
+class FailedChargeCalculator(
+    private val failedChargeCounterManager: FailedChargeCounterManager,
+    private val eventStore: EventStore
+) {
+    fun calculateAndUpdateNumberOfFailedCharges(memberId: String, timestamp: Instant): FailedChargeCounter {
+        val numberFailedCharges = EventStoreUtils.numberOfLastContinuousFailedCharges(memberId, eventStore)
+        failedChargeCounterManager.updateNumberFailedCharges(memberId, numberFailedCharges, timestamp)
+
+        return failedChargeCounterManager.getFailedChargeCounter(memberId)!!
```
johantjellden, line 17: Isnt numberFailedCharges == failedChargeCounterManager.getFailedChargeCounter(memberId)!!
If so why not return it directly?

PR #89, opened 2020-07-01
fixing bug with failed charges

meletios:
Bug : We are notifying customers that they have a failed charge even if they are terminated previously for other reasons.

Fix: Dont send failed charge event to customer io if member is already terminated before!


src/main/kotlin/com/hedvig/accountservice/termination/TerminationServiceImpl.kt
```diff
@@ -70,14 +76,7 @@ class TerminationServiceImpl(
...
-                        memberId = memberId,
-                        numberOfFailedCharges = memberTerminationOutcome.numberOfFailedCharges
-                    )
-                } else {
-                    memberTerminationOutcome
-                }
+                memberTerminationOutcome
             }
             is MemberTerminationOutcome.NotTerminated -> memberTerminationOutcome
             is MemberTerminationOutcome.NoContractsToTerminate -> throw IllegalStateException("Already handled in the MemberTerminationOutcome.Terminated flow")
```
FredrikAreschoug, line 82:
```suggestion
            is MemberTerminationOutcome.NoContractsToTerminate -> throw IllegalStateException("Already handled in the MemberTerminationOutcome.")
```

meletios: Goood point, will fix

PR #95, opened 2020-09-18
do not terminate norwegian members this month

AlexSJ1990:
Stops us terminating Norwegian members this month
removes early return from terminateContract function (this was put in as a temporary fix last month to avoid terminating all members)


meletios (approved):
💯

PR #96, opened 2020-10-15
if market is Norway require 3 failed charges before termination, if s…

AlexSJ1990:
…weden 2 failed charges and if no market can be found 3 failed charges

This is a bit of a hack for this month -> from next month onwards we will terminate after 2 failed charges and so will remove this function.


FredrikAreschoug (dismissed):
Nice! If we want to be super thorough we could test the exception path but that is just a suggestion you do as you please :slightly_smiling_face:

FredrikAreschoug (approved):
amaze!

PR #98, opened 2020-10-16
add index to speed up scheduler states query

AlexSJ1990:


FredrikAreschoug (approved):

src/main/resources/db/changelog/migrations/20201016-1111-add-scheduler-states-member-id-created-at-index.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset alex:20201016-1111-add-scheduler-states-member-id-created-at-index
+
+CREATE INDEX scheduler_states_member_id_changed_at_idx ON scheduler_states(member_id, changed_at desc)
```
vonElfvin, line 5: Why idx?
```suggestion
CREATE INDEX scheduler_states_member_id_changed_at ON scheduler_states(member_id, changed_at desc)
```

vonElfvin: image
The doesn't look the same

vonElfvin: I think `CONCURRENTLY` is good to have as well

johantjellden: With regards to the "idx" postfix.
We dont have a naming standard for our databases yet. For this case look at the other indexes in account-service and make sure we follow that standard.

johantjellden (approved):

PR #107, opened 2020-10-27
First stab at making failed charges distinct per month

vonElfvin:
Makes charges distinct per period, where a charge is attributed to the period before the 5th of the coming period


AlexSJ1990 (approved):
👍

PR #110, opened 2020-10-28
Return correct charge date in the apps (i.e. when the charge went thr…

vonElfvin:
…ough and not when it was started)


src/main/kotlin/com/hedvig/accountservice/graphql/resolvers/Query.kt
```diff
@@ -48,7 +50,7 @@ class Query(
         return chargeEntries.map { entry ->
             Charge(
                 amount = MonetaryAmountV2.of(entry.amount.negate()),
-                date = entry.fromDate
+                date = LocalDate.ofInstant(entry.chargedAt!!, STOCKHOLM_ZONE_ID)
```
vonElfvin, line 53: Note the filter above ☝️

meletios (approved):

PR #111, opened 2020-11-03
add new state if we are unable to charge members

AlexSJ1990:
if we fail to get an answer back from payment-service when initiating the charge, move member to ChargeRequestError state. We can then reset these members and try to charge them again.


FredrikAreschoug (commented):
Looks good! But should we add a reason and the exception from why we were unable to charge? I would like it when debugging :slightly_smiling_face:

johantjellden:
> Looks good! But should we add a reason and the exception from why we were unable to charge? I would like it when debugging 🙂

Yes and lets not consider this done until we have code in place that dispatches the new command.

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -389,6 +395,24 @@ class MemberScheduler {
...
+        )
+
+        apply(ChargeUnableToBeInitiatedEvent(command.memberId, nextState))
+        logger.error { "Unable to initiate charge for memberId=${command.memberId}" }
+    }
+
+    @EventSourcingHandler
+    fun on(event: ChargeUnableToBeInitiatedEvent) {
+        currentState = event.nextState
+    }
```
vonElfvin, line 414: Let's make sure this is reflected in the database

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -59,12 +59,17 @@ class MemberScheduler {
...
+            ChargeStatus.UNABLE_TO_INITIATE_CHARGE
         ),
         ChargeStatus.CHARGE_REQUEST_FAILED to listOf(
             ChargeStatus.WAITING_FOR_SUBSCRIPTION,
             ChargeStatus.BLACKLISTED
         ),
+        ChargeStatus.UNABLE_TO_INITIATE_CHARGE to listOf(
+            ChargeStatus.WAITING_FOR_SUBSCRIPTION,
+            ChargeStatus.BLACKLISTED
+        ),
```
vonElfvin, line 72: Should we add "CHARGE_REQUESTING" here, i.e. that you can try again without going through the whole loop

src/main/kotlin/com/hedvig/accountservice/schedulers/model/ChargeStatus.kt
```diff
@@ -11,6 +11,7 @@ enum class ChargeStatus {
     SCHEDULING_CHARGE,
     CHARGE_REQUESTING,
     CHARGE_REQUEST_FAILED,
+    UNABLE_TO_INITIATE_CHARGE,
```
vonElfvin, line 14:
```suggestion
    UNABLE_TO_REQUEST_CHARGE,
```

vonElfvin: Maybe
```suggestion
    CHARGE_REQUEST_ERROR,
```

vonElfvin (changes requested):
Great job, clean solution. Missing @EventListener however for the database

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
AlexSJ1990: yess very good idea, I was wondering if this was okay to do!

src/test/kotlin/com/hedvig/accountservice/schedulers/domain/MemberSchedulerTest.kt
```diff
@@ -581,6 +582,71 @@ class MemberSchedulerTest {
             }
     }
 
+    @Test
```
vonElfvin, line 585: Should probably add a test to the `ChargeSaga` as well so we know it sends this command when it gets and exception in the product-pricing client

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/commands/UnableToRequestChargeCommand.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import org.axonframework.commandhandling.TargetAggregateIdentifier
+import javax.money.MonetaryAmount
+
+data class UnableToRequestChargeCommand(
+    @TargetAggregateIdentifier
+    val memberId: String,
+    val amount: MonetaryAmount,
+    val scheduledBy: String
+)
```
vonElfvin, line 11: If you want to log it
```suggestion
data class UnableToRequestChargeCommand(
    @TargetAggregateIdentifier
    val memberId: String,
    val amount: MonetaryAmount,
    val scheduledBy: String,
    val exception: Exception
)
```

src/main/kotlin/com/hedvig/accountservice/schedulers/domain/MemberScheduler.kt
```diff
@@ -389,6 +396,24 @@ class MemberScheduler {
...
+    fun handle(command: UnableToRequestChargeCommand) {
+        val nextState = SchedulerState(
+            memberId = command.memberId,
+            status = ChargeStatus.CHARGE_REQUEST_ERROR,
+            changedBy = command.scheduledBy,
+            amount = command.amount
+        )
+
+        apply(ChargeRequestErrorEvent(command.memberId, nextState))
+        logger.error { "Unable to request charge for memberId=${command.memberId}, exception: ${command.exception}" }
```
vonElfvin, line 409: Aparently you can do
```suggestion
        logger.error (command.exception) { "Unable to request charge for memberId=${command.memberId}" }
```

FredrikAreschoug (approved):

vonElfvin (approved):

